PR #13, opened 2014-06-05
Ensure strictness in the state in modify'.

kantp:
The old implementation of modify' used state f with a function f that was strict in the new state.  However, in state itself, f is applied to the state in a non-strict way, so the strictness is lost.

The implementation suggested here should ensure that the state is evaluated.


RyanGlScott:
Well spotted. I've merged this, and tweaked the implementation to be more similar to the one in `transformers`.

PR #37, opened 2016-07-08
Added `liftEither` for lifting Either into MonadError

ElvishJerricco:
Lifts an `Either e` into any `MonadError e`.

``` haskell
do { val <- liftEither =<< action1; action2 }
```

where `action1` returns an `Either` to represent errors.


ElvishJerricco:
@ekmett, in hindsight, perhaps this should have been called `throwEither`, and accompanied by `throwMaybe :: MonadError e m => e -> Maybe a -> m a`. What do you think?

ocharles:
I think `lift` is fine, but would also like `liftMaybe`.

PR #43, opened 2017-07-18
Modernize .cabal file & update Travis CI job

hvr:
This updates the .cabal format to cabal-version:1.10
and turns `extensions:` into per-module pragmas
(this makes it easier to integrate `mtl` into GHC's
cabal-less `ghc-cabal` bootstrapping phase)

Note, this implicitly drops support for GHC < 7 as this sets the
requirement on a compiler that supports Haskell2010.


hvr:
@ekmett note that I don't need an actual release for GHC's sake just yet; this can happily stay in `master` until GHC 8.4.1 or something else forces a new release of  `mtl`

PR #46, opened 2017-08-29
Add MonadWriter instance for bare tuples

Taneb:
This seems a natural instance to have, comparable to the ((->) r) MonadReader instance


ekmett:
We have the Monad instance.

PR #49, opened 2017-11-20
Put a warning about how ExceptT is different from Control.Exception

saurabhnanda:
We have been bitten by this confusion first-hand and would like to prevent others from having this experience.

### DIFF
--- a/Control/Monad/Except.hs
+++ b/Control/Monad/Except.hs
@@ -31,6 +31,8 @@ The Error monad (also called the Exception monad).
 -}
 module Control.Monad.Except
   (
+    -- * Warning
+    -- $warning
     -- * Monads with error handling
     MonadError(..),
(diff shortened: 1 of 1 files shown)


Control/Monad/Except.hs
```diff
@@ -71,6 +73,18 @@ import Control.Monad.Fix
 import Control.Monad.Instances ()
 #endif
 
+{- $warning
+Please do not confuse 'ExceptT' and 'throwError' with 'Contorl.Exception.Exception' /
```
cocreature: Typo: `Contorl` instead of `Control`.

ekmett: Fixed in c3078d

saurabhnanda:
@cocreature @ekmett that was fast 😄 Thank you 👍

Issue #44, opened 2017-07-24
No instances for recently-added monads from transformers


bucklereed:
Control.Monad.Trans.Accum was added in transformers 0.5.3.0, but currently it's slightly painful to use because it doesn't have the usual MonadState, MonadReader, etc instances yet.

Whether it should get a MonadWriter instance that lifts the operations to the inner monad or handles them itself is a question.

SelectT was also added in transformers 0.5.3.0 and similarly lacks instances.

RyanGlScott:
We should definitely add instances for `AccumT` and `SelectT`.

As far as how `AccumT`'s instances should work, I'm not sure. @fizruk had proposed in #32 that we introduce a new type class for things that can support `AccumT`'s operations (`MonadAccum`?), but it's not clear to me what its relationship with `MonadWriter`/`MonadState` would be.

I think there are enough design questions here that it would be beneficial to draft up a sketch of how this would work and propose it to the [Haskell libraries mailing list](https://mail.haskell.org/mailman/listinfo/libraries) for feedback.

fizruk:
I think that `AccumT` should propagate `MonadReader`, `MonadWriter` and `MonadState` to inner monad and that we should introduce `MonadAccum`. This would be consistent with how other classes and transformers work.

RyanGlScott:
@fizruk, that sounds reasonable to me. But the design of these `Monad-` classes are usually more involved than one would initially believe (at least in my experience), so I think it would help to come up with a concrete design and propose it to the mailing list. If the community likes it, I'll accept a PR which implements it.

ekmett:
I'm also leaning towards adding MonadAccum and the like.

treeowl:
I just got bitten by the lack of a `MonadError` instance for `AccumT`.

minad:
See also #38

chessai:
This is resolved in master, pending 2.3 release

PR #51, opened 2018-01-13
MonadAccum, MonadSelect, + mtl instances for SelectT/AccumT

mstksg:
Can't seem to find a branches with progress on #44, so here's my shot.

Tried my best to follow the structure of the other modules.  WIP, as I've only added MonadAccum and mtl instances for AccumT.

Some notes --

* For instances of MonadCont and MonadError and MonadWriter, etc., with somewhat "involved" functions (callCC, catch, listen, pass), the convention seems to be to implement these lifted functions in *transformers* (`liftListen`, etc.) instead of mtl.  It would probably make more sense to do the same for `AccumT` (define `Control.Monad.Trans.Accum.liftListen` in *transformers*, etc.) than in here, so we might want to get that merged into *transformers* before committing to the currently inlined defined versions I have here.

* `MonadAccum` only has `add` and `look`.  Having it include `accum`, I feel, (like `MonadState` has `state`) would be pretty strongly against the spirit of the interface.

* Not exactly sure what to put for copyrights on these new modules?

### DIFF
--- a/Control/Monad/Accum.hs
+++ b/Control/Monad/Accum.hs
@@ -1,0 +1,54 @@
+{-# LANGUAGE CPP #-}
+{- |
+Module      :  Control.Monad.Accum
+Copyright   :  ???
+License     :  BSD-style (see the file LICENSE)
+
+Maintainer  :  libraries@haskell.org
+Stability   :  experimental
+Portability :  non-portable (multi-parameter type classes)
+
+Provides append-only accumulation during the computation.  For more general
+access, use "Control.Monad.State" instead.
+-}
+
+{-
+  Rendered by Michael Weber <mailto:michael.weber@post.rwth-aachen.de>,
+  inspired by the Haskell Monad Template Library from
+    Andy Gill (<http://web.cecs.pdx.edu/~andy/>)
+-}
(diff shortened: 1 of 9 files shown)


RyanGlScott:
Hi @mstksg, my sincerest apologies for not looking at this sooner.

I'll preface this comment by saying that this is a pretty sizable addition to `mtl`, and moreover, one that others will likely have feedback on. I think it might be wise to prepare a small document stating:

1. Exactly which classes you are proposing to add, and what the types of their methods will be.
2. The implementations of non-trivial instances, such as the `MonadWriter` instance for `AccumT`, stating the rationale behind it.

Echoing my comments in https://github.com/haskell/mtl/issues/44#issuecomment-317767868, it would be wise to run this document past the [libraries mailing list](https://mail.haskell.org/mailman/listinfo/libraries) once to make sure folks are on board with it. Once we've agreed on a design, we can proceed.

> Not exactly sure what to put for copyrights on these new modules?

Well, you created them—put yourself and the current year!

turion:
What's the progress on this?

chessai:
closing in favour of #80

PR #53, opened 2018-03-28
Add imports to Control.Monad.Reader examples and other small tweaks

rcook:
It's a common use case for readers of the documentation to copy and paste the code samples straight into an editor. This changes makes sure that these code samples compile with no warnings.

### DIFF
--- a/Control/Monad/Reader.hs
+++ b/Control/Monad/Reader.hs
@@ -80,7 +80,11 @@ The variable @count@ contains number of variables in the bindings.
 You can see how to run a Reader monad and retrieve data from it
 with 'runReader', how to access the Reader data with 'ask' and 'asks'.
 
-> type Bindings = Map String Int;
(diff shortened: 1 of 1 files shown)


RyanGlScott:
Thanks @rcook, that definitely seems like an improvement.

One thing I'm iffy on is the use of `HashMap` instead of `Map`—only because it uses something from an external library, `unordered-containers`. Granted, `mtl` depends on neither `containers` nor `unordered-containers`, so using `Data.Map` from `containers` is technically also "external", but `containers` does have the benefit of being bundled with GHC itself, so there's it's less likely to cause problems for anyone who copies-and-pastes this code. Do you agree?

rcook:
@RyanGlScott : Yes, that makes sense. I've updated the commit to use `Data.Map` instead of `Data.HashMap.Strict`. Thanks!

PR #54, opened 2018-04-21
Add a convenience function withEnv

marmistrz:
It's useful when you're storing the environment in a data structure.

### DIFF
--- a/Control/Monad/Reader/Class.hs
+++ b/Control/Monad/Reader/Class.hs
@@ -43,7 +43,7 @@ than using the 'Control.Monad.State.State' monad.
 
 module Control.Monad.Reader.Class (
     MonadReader(..),
-    asks,
+    asks, withEnv
     ) where
 
 import Control.Monad.Trans.Cont as Cont
@@ -97,6 +97,10 @@ asks :: MonadReader r m
     => (r -> a) -- ^ The selector function to apply to the environment.
     -> m a
 asks = reader
+
+-- | Executes a computation in a predefined environment.
+withEnv :: MonadReader r m => r -> m a -> m a
+withEnv env = local (const env)
 
 -- ----------------------------------------------------------------------------
 -- The partially applied function type is a simple reader monad


chessai:
I agree that this seems useful. Perhaps others might as well, and chime in.

Can you add a changelog entry?

andrewthad:
I agree that such a function is useful. However, if it's going be in `mtl`, I would rather see its monomorphic cousin appear in `transformers` first. The name feels a little disconnected. The function is closely related to `local`, and none of the existing reader functions mention `with` or `Env` in their name. A few ideas (not that I particularly like any of these): `localEnv`, `localConst`, `local'` (gross), `local_`.

chessai:
> I agree that such a function is useful. However, if it's going be in `mtl`, I would rather see its monomorphic cousin appear in `transformers` first. The name feels a little disconnected. The function is closely related to `local`, and none of the existing reader functions mention `with` or `Env` in their name. A few ideas (not that I particularly like any of these): `localEnv`, `localConst`, `local'` (gross), `local_`.

This makes sense. An issue should first be opened on transformers then.

chessai:
btw, perhaps `readerM` is a good name? because we have

```haskell
reader :: MonadReader r m => (r -> a) -> m a
```

so `readerM` makes a nice contrast, just indicating that the function is monadic:

```haskell
readerM :: (r -> m a) -> m a
readerM f = ask >>= f
```

chessai:
> btw, perhaps `readerM` is a good name? because we have
> 
> ```haskell
> reader :: MonadReader r m => (r -> a) -> m a
> ```
> 
> so `readerM` makes a nice contrast, just indicating that the function is monadic:
> 
> ```haskell
> readerM :: (r -> m a) -> m a
> readerM f = ask >>= f
> ```

I had something else in mind when writting this comment. disregard.

sjshuck:
`withEnv` seems like the wrong name; we don't call anything "env" in the API.  I would go with `local_`.

It's trivial to implement as `local . const` and there's been no activity for 5 years so I'm closing.  Feel free to open an issue in https://github.com/haskell/transformers if still desired.

PR #58, opened 2018-07-04
Try out omit-local-package

phadej:


RyanGlScott:
I believe I tried this in https://github.com/haskell/mtl/commit/1c110d117d4e5a554d96a39879c14791696c634d, but it still failed to build on GHC 7.8: https://travis-ci.org/haskell/mtl/builds/399714580

phadej:
Oh, I see.

phadej:
So `mtl` as in master is not buildable with installed version of `transformers`, And IIRC that's intentiona? So we could have `UNCONSTRAINED...` pushed new commit, let's see.

RyanGlScott:
Good point, I didn't even notice the lower version bounds on `transformers`... that would definitely explain that.

Maybe this isn't the best place to try out your new `haskell-ci` feature, then. `unix` and `stm` experienced similar issues, so perhaps those would be better testing grounds.

PR #61, opened 2019-03-03
Add laws for MonadState and MonadReader

Lysxia:
See #5. These two classes are the easiest to think about.


Lysxia:
Ping @RyanGlScott

RyanGlScott:
I really don't feel qualified to make a judgment about whether these should be _the_ laws for these classes, which is why I haven't taken action. It might be worthwhile to bring up this topic on the Haskell libraries mailing list in order to garner a consensus (which I'd find convincing).

Lysxia:
I see, thanks for your input! I will start a discussion on the mailing list.

Lysxia:
This is still WIP. [On the mailing list](https://mail.haskell.org/pipermail/libraries/2019-April/029553.html) it was suggested to investigate the matter more formally. So I started studying those laws in Coq: https://github.com/Lysxia/coq-mtl but I've been distracted by other things lately.

The situation, as I understand it, is that there is no formal definition of the mtl classes, and there is no objective criterion for what makes a law "good" or "bad". Through the formalization, the hope is to get a better understanding of how the laws relate to each other, and ideally to tease out some nice structure that could convincingly be considered the "essence" of mtl.

hanshoglund:
How about adopting the uncontroversial idea that `reader`, `writer` and `state` are monad homomorphisms as a starting point?

Lysxia:
It's been a week since I reiterated my proposal: https://mail.haskell.org/pipermail/libraries/2019-October/030038.html

Is there anything more I can do to make progress?

PR #62, opened 2019-03-04
Add laws for MonadError and add tryError

Lysxia:
`tryError` is used to state an additional law about the interaction of `catchError` with `(>>=)`.

For reference, the 5 laws other than that one can be found in the paper *Just `do` it: Simple Monadic Equational Reasoning*, Section 5 (modulo a change from `catch :: m a -> m a -> m a` to `catchError :: m a -> (e -> m a) -> m a`).


Lysxia:
I also cut the original comment on this class. I find the first sentence quite hard to read, and the rest is not quite relevant.

> It is common to use `Either String`

IMO `Either` is a very niche instance of `MonadError`.

> You can also define your own error type and\/or use a monad type constructor
> other than @'Either' 'String'@ or @'Either' 'IOError'@.
￼> In these cases you will have to explicitly define instances of the 'MonadError'
￼> class.

You don't need to define additional instances if you just want to vary your exception type from `String` or `IOError` (I would assume this is a remnant of the old `MonadError`).

> (If you are using the deprecated "Control.Monad.Error" or "Control.Monad.Trans.Error", you may also have to define an 'Error' instance.)

The old `MonadError` has been deprecated for so long, I would assume people who have to maintain such old code should be able to follow the changelog.

Lysxia:
See #5 . With #61, the remaining effects to document are `Writer` and `Cont`.

Lysxia:
Please also  note that everything I'm doing here is open for questions and discussion.

Lysxia:
This probably needs a `@since` annotation but I don't know what version.

PR #65, opened 2019-06-23
Export evalCont/evalContT

### LINKED ISSUE #64: Re-export evalCont/evalContT?
These two functions are present in `transformers` but not `mtl`:

* `evalCont :: Cont r r -> r` -- The result of running a CPS computation with the identity as the final continuation.
* `evalContT :: Monad m => ContT r m r -> m r` -- The result of running a CPS computation with return as the final continuation.

Should they be re-exported? If so, I will prepare a PR.

endgame:
Fixes #64, and includes a version bump and changelog entry in a separate commit, in case you want to cherry-pick one or the other.

### DIFF
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -1,3 +1,7 @@
+Unreleased
+----------
+* `Control.Monad.Cont` now re-exports `evalCont` and `evalContT`
+
 2.2.2
 -----
 * `Control.Monad.Identity` now re-exports `Control.Monad.Trans.Identity`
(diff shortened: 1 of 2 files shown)


chessai:
Thanks! Not sure we should include a version bump on this commit though. I think we typically make a commit doing that prior to a release.

endgame:
I have force-pushed a single commit that does the exports and a changelog entry (under "unreleased").

chessai:
No need to force push. We'd just squash the commits. Thanks for the thoughtfulness about the tree, though.

Anyway, LGTM. Thanks!

PR #66, opened 2019-06-24
add tryError, withError, handleError, mapError (#60)

### LINKED ISSUE #60: Suggested addition tryError (originally tryExceptT)
It seems to me that this function might make a useful addition to Control.Monad.Except:

    -- | ExceptT analog to the 'try' function.
    tryExceptT :: Monad m  => ExceptT e m a -> ExceptT e m (Either e a)
    tryExceptT = lift . runExceptT

ddssff:
Fixes #60

### DIFF
--- a/Control/Monad/Error/Class.hs
+++ b/Control/Monad/Error/Class.hs
@@ -48,6 +48,7 @@ module Control.Monad.Error.Class (
     mapError,
   ) where
 
+import qualified Control.Exception
 import Control.Monad.Trans.Except (Except, ExceptT)
 import Control.Monad.Trans.Error (Error(..), ErrorT)
 import qualified Control.Monad.Trans.Except as ExceptT (throwE, catchE)
@@ -195,7 +196,7 @@ instance (Monoid w, MonadError e m) => MonadError e (StrictWriter.WriterT w m) w
     throwError = lift . throwError
     catchError = StrictWriter.liftCatch catchError
 
--- | MonadError analog to the 'try' function.
+-- | 'MonadError' analog to the 'Control.Exception.try' function.
 tryError :: MonadError e m => m a -> m (Either e a)
 tryError action = (liftM Right action) `catchError` (return . Left)
 
@@ -205,10 +206,13 @@ tryError action = (liftM Right action) `catchError` (return . Left)
 withError :: MonadError e m => (e -> e) -> m a -> m a
 withError f action = tryError action >>= either (throwError . f) return
 
--- | As 'handle' is flipped 'catch', 'handleError' is flipped 'catchError'.
+-- | As 'handle' is flipped 'Control.Exception.catch', 'handleError'
+-- is flipped 'catchError'.
 handleError :: MonadError e m => (e -> m a) -> m a -> m a
 handleError = flip catchError
 
--- | MonadError analogue of the 'mapExceptT' function.
+-- | 'MonadError' analogue of the 'mapExceptT' function.  The
+-- computation is unwrapped, a function is applied to the @Either@, and
+-- the result is lifted into the second 'MonadError' instance.
 mapError :: (MonadError e m, MonadError e' n) => (m (Either e a) -> n (Either e' b)) -> m a -> n b
 mapError f action = f (tryError action) >>= liftEither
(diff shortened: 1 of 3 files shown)


chessai:
This LGTM. @RyanGlScott any thoughts?

chessai:
Oh, needs a changelog entry!

ddssff:
I see there is another PR with tryError.

Lysxia:
> I see there is another PR with tryError.

No worries about that. That PR I wrote is really meant to do something else, so I can rebase it on top of this later. I thought I needed `tryError` to write some property but actually I don't anymore (because the property is invalid).

chessai:
@ddssff the builds failed because older versions of base's Prelude do not export `<$>`. i think on all GHCs before 7.10.3 you need to `import Data.Functor ((<$>))`

ddssff:
MIN_VERSION_base(4,8,0) if I'm not mistaken

chessai:
yeah, something like

```haskell
#if !MIN_VERSION_base(4,8,0)
import Data.Functor ((<$>))
#endif
```

Alternatively, we could just use `fmap`, which has been exported by the Prelude for all supported versions of base. This would remove that bit of CPP. I don't have strong opinions about that, though I do prefer the least amount of CPP possible.

ddssff:
Oh frabjous day.

Control/Monad/Error/Class.hs
```diff
@@ -190,3 +194,21 @@ instance (Monoid w, MonadError e m) => MonadError e (LazyWriter.WriterT w m) whe
 instance (Monoid w, MonadError e m) => MonadError e (StrictWriter.WriterT w m) where
     throwError = lift . throwError
     catchError = StrictWriter.liftCatch catchError
+
+-- | MonadError analog to the 'try' function.
```
RyanGlScott: I don't think `'try'` will be hyperlinked in the rendered Haddocks, since `Control.Exception` doesn't appear to be imported. I think the usual workaround for this problem is to `import qualified Control.Exception` and use `'Control.Exception.try'` in the Haddocks.

Control/Monad/Error/Class.hs
```diff
@@ -190,3 +194,21 @@ instance (Monoid w, MonadError e m) => MonadError e (LazyWriter.WriterT w m) whe
...
+tryError :: MonadError e m => m a -> m (Either e a)
+tryError action = (liftM Right action) `catchError` (return . Left)
+
+-- | Modify the value (but not the type) of an error.  The type is
```
RyanGlScott: It might be worth mentioning the very similar `withExcept` (or `withExceptT`) function here.

Control/Monad/Error/Class.hs
```diff
@@ -190,3 +194,21 @@ instance (Monoid w, MonadError e m) => MonadError e (LazyWriter.WriterT w m) whe
...
+-- fixed because of the functional dependency @m -> e@.  If you need
+-- to change the type of @e@ use 'mapError'.
+withError :: MonadError e m => (e -> e) -> m a -> m a
+withError f action = tryError action >>= either (throwError . f) return
+
+-- | As 'handle' is flipped 'catch', 'handleError' is flipped 'catchError'.
```
RyanGlScott: Similarly, use `'Control.Exception.catch'` in the Haddocks.

Control/Monad/Error/Class.hs
```diff
@@ -190,3 +194,21 @@ instance (Monoid w, MonadError e m) => MonadError e (LazyWriter.WriterT w m) whe
...
+handleError :: MonadError e m => (e -> m a) -> m a -> m a
+handleError = flip catchError
+
+-- | MonadError analogue of the 'mapExceptT' function.
```
RyanGlScott: It might be wise to recap what this function does (e.g., "Map the unwrapped computation using the given function").

Also, s/`MonadError`/`'MonadError'`/

chessai: resolved

Control/Monad/Error/Class.hs
chessai: resolved

Control/Monad/Error/Class.hs
chessai: resolved

Control/Monad/Error/Class.hs
chessai: resolved

chessai:
Thanks!

Issue #38, opened 2016-09-01
Stricter Writer


minad:
Hi,

I made the packages writer-cps-mtl and writer-transformers-mtl which implement a stricter writer monad transformer in continuation passing style as was discussed before on the Haskell mailing list etc
- http://hackage.haskell.org/package/writer-cps-mtl
- http://hackage.haskell.org/package/writer-cps-transformers

See also the current reddit discussion:
https://www.reddit.com/r/haskell/comments/50hf2s/stricter_writert_and_rwst_replacements/

Would you be interested in including this into mtl? If not it can also exist as a separate package, but I guess it would be nice to have a well-behaving writer around in the standard toolbox.

Daniel

int-index:
@minad Thank you for `writer-cps-*` packages, I'll incorporate them into the next release of `ether` as the default implementation for `Writer`.

minad:
It seems transformers pulled writer-cps-transformers, see https://hub.darcs.net/ross/transformers/patch/b6743fad9f5ce8a9e13f15320f7d2752330084e2

RyanGlScott:
Assuming that these provide the same interface as the other `State`/`Writer` data types (which appears to be the case from a cursory glance, although I have not looked deeply), I'd happily accept a patch which adds `MonadState`/`MonadWriter` instances for these.

minad:
As soon as the new transformers is released, you can pull the code from http://hackage.haskell.org/package/writer-cps-mtl. I can also prepare a patch. The same applies to the other writer-cps-* packages, providing orphan instances for exceptions, lens and mmorph.

RyanGlScott:
I believe #67 implements this on `mtl`'s side, so closing. Please re-open if you disagree.

PR #67, opened 2019-07-24
Support Control.Monad.Trans.{Writer,RWS}.CPS from transformers 0.5.6

lexi-lambda:
`transformers-0.5.6.0` adds new `Control.Monad.Trans.Writer.CPS` and `Control.Monad.Trans.RWS.CPS` monad transformers, so this PR adds support for them in `mtl`.

I’ve taken the liberty of bumping the version to 2.3, since this seems like a significant enough change to do so, and picking what the next release’s version number will be was necessary for the `@since` annotations in the docs. Let me know if you’d like me to do something else, and I’ll change it.

### DIFF
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -1,5 +1,6 @@
 Unreleased
 ----------
+* Add instances for `Control.Monad.Trans.Writer.CPS` and `Control.Monad.Trans.RWS.CPS` from `transformers` 0.5.6 and add `Control.Monad.Writer.CPS` and `Control.Monad.RWS.CPS`.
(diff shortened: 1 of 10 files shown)


fishtreesugar:
It could fix #38

chessai:
At a glance, this looks great! But could you remove the version bump? That's something that's typically done prior to a release, and not based on any one contribution alone.

lexi-lambda:
@chessai Sure, I’ve removed the version bump. I’ve left the Haddock annotations alone, so they still say `@since 2.3`, since I’m not sure what else to do with those. Let me know if there’s something better.

RyanGlScott:
Thanks!

PR #70, opened 2020-01-18
Switch to MonadReader based MonadState implementation for ContT

Atry:
The implementation is inspired by https://github.com/ekmett/kan-extensions/blob/f9f8717883d250c2067c13b1e867274c29d7095e/src/Control/Monad/Codensity.hs#L251

### DIFF
--- a/Control/Monad/State/Class.hs
+++ b/Control/Monad/State/Class.hs
@@ -51,6 +51,7 @@ import qualified Control.Monad.Trans.RWS.CPS as CPSRWS (RWST, get, put, state)
 import Control.Monad.Trans.Writer.CPS as CPS
 #endif
 
+import qualified Control.Monad.Reader.Class as MonadReader
 import Control.Monad.Trans.Class (lift)
 import Control.Monad
 import Data.Monoid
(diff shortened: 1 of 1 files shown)


chessai:
This is interesting, but I'm not sure if it's at all warranted. Especially considering that it will cause breakage. I'm leaning toward a no on this.

cartazio:
Could you explain the motivation and context?

Either way thanks for sharing. But some exposition About  context and such
would be appreciated

Atry:
The problem is that the existing `instance MonadState s m => MonadState s (ContT r m)` conflicts with reader based instance.

chessai:
Could you elaborate?

Atry:
@cartazio 
A `ReaderT` has a simpler implementation than `StateT`. Suppose we have the following type:
``` Haskell
type X = ContT Text (ReaderT String (ReaderT Double)
```
With the help of this PR, `X` will support both `MonadReader String X` and `MonadState String X`, and with the help of https://github.com/haskell/mtl/pull/73, we can even have `MonadReader Double X` and `MonadState Double X`. Therefore, monad transformers can be easily extended, used as a simple and fast alternative to `Freer` / `Eff` things.

Atry:
The trick to implement `MonadState` from `MonadReader` is explored by @ekmett 9 years ago, this PR just applies the same idea to `ConT`

chessai:
This would be a weird and confusing departure from the typical lifting witnessed by monad transformers. That is, MTL instances typically lift like:

```haskell
instance C x m => C x (t m)
```

and not

```haskell
instance C0 x m => C1 x (t m)
```

While clever and certainly used by Codensity, this would cause a lot of breakage and confusion. I'm a strong -1 on this. Feel free to re-open if you want to make a case for this being changed.

cartazio:
agreed

PR #72, opened 2020-01-18
Add MonadWriter instance for ContT

Atry:
Similar to https://github.com/haskell/mtl/pull/70

### DIFF
--- a/Control/Monad/Writer/Class.hs
+++ b/Control/Monad/Writer/Class.hs
@@ -30,6 +30,7 @@ module Control.Monad.Writer.Class (
     censor,
   ) where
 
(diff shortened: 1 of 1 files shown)


chessai:
Also similarly to #70, why not using the underlying `MonadWriter w m` instance?

chessai:
(The above question was rhetorical, you can't.)

But I'm not sure that we do want to lift reader in this way? Thoughts @haskell/core-libraries-committee ?

gwils:
This instance seems weird to me. Besides which it would be confusing for an MTL class to have an instance that wasn't the expected lifting.

chessai:
non-canonical, closing

Issue #75, opened 2020-06-06
Plans for ListT and ErrorT


fumieval:
They are finally removed after several years of deprecation (https://hub.darcs.net/ross/transformers/patch/7c809b6f9db019ed761c971d10c86fa004f60d89). Is it OK to simply remove them from `mtl` as well? If so I can submit a PR.

chessai:
Yes. I would accept a PR for this. Make sure you take care with the
relevant CPP.

fumieval:
Created https://github.com/haskell/mtl/pull/76 

> Make sure you take care with the relevant CPP

Can you elaborate a bit? No existing CPP seems to be affected

PR #76, opened 2020-06-07
Remove ListT/ErrorT-related stuff

fumieval:
They are removed from transformers; we should remove them from mtl too (#75)


chessai:
@fumieval the CPP i was referring to in #75 was for allowing for these instances with older transformers. But, I'm tempted to say that these can just be removed in the next majour version bump, without the need for CPP.

cabal.project
```diff
@@ -1 +1,2 @@
 packages: .
+  transformers/
```
chessai, line 2:
```suggestion
\n
```

cabal.project
```diff
@@ -1 +1,2 @@
 packages: .
+\n
```
chessai, line 2:
```suggestion
```

chessai:
I am confused as to why github is showing that "All checks have failed", because all of them succeeded, except for HEAD. Perhaps it is to do with the strange HEAD failure, which maybe @RyanGlScott can make light of: https://travis-ci.org/github/haskell/mtl/jobs/707484624

RyanGlScott:
I'm not sure what to make of HEAD-related failures either. But for what it's worth, the version of GHC HEAD that is provided at https://launchpad.net/~hvr/+archive/ubuntu/ghc (which `haskell-ci` uses) is quite old (around GHC 8.7 or so),. I've opted to remove the `ghc-head` configurations in all of my `haskell-ci` configs, since a HEAD that old is basically useless.

chessai:
@RyanGlScott thanks, I'll probably just remove it then.

chessai:
Can't get CI for the old HEAD to pass, so I'm going to just disregard it. HEAD has been removed from the .travis.yml on master.

PR #80, opened 2020-07-13
add (existing mtl typeclass) instances for AccumT

chessai:
Like it says on the tin.

Note that this PR intentionally DOES NOT address adding MonadAccum, since IMO that is a larger concern and should be handled separately.

### DIFF
--- a/Control/Monad/Writer/Class.hs
+++ b/Control/Monad/Writer/Class.hs
@@ -211,7 +211,13 @@ instance MonadWriter w m => MonadWriter w (Strict.StateT s m) where
 
 #if MIN_VERSION_transformers(0,5,3)
 -- | @since 2.3
-instance (Monoid w, MonadWriter w m) => MonadWriter w (AccumT w m) where
+instance
+  ( Monoid w
+  , MonadWriter w m
+#if !MIN_VERSION_base(4,8,0)
+  , Functor m
+#endif
+  ) => MonadWriter w (AccumT w m) where
     writer = lift . writer
     tell   = lift . tell
     listen = Accum.liftListen listen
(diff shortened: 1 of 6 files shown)


chessai:
Requesting review from @ekmett @RyanGlScott and anyone who is interested from the CLC, cc @haskell/core-libraries-committee

Control/Monad/Writer/Class.hs
```diff
@@ -204,3 +208,18 @@ instance MonadWriter w m => MonadWriter w (Strict.StateT s m) where
...
+
+#if MIN_VERSION_transformers(0,5,3)
+-- | @since 2.3
+instance
+  ( Monoid w
+  , MonadWriter w m
+#if !MIN_VERSION_base(4,8,0)
+  , Functor m
+#endif
+  ) => MonadWriter w (AccumT w m) where
```
RyanGlScott, line 220: Per the discussion in #44, there is a design choice to be made about how the `MonadWriter` instance for `AccumT` works. It could either:

1. Lift the operations to the inner monad, or
2. Handle the operations itself.

This PR chooses approach (1), which I think is fine. But it a choice nonetheless, and I think it should be documented to avoid any confusion on the topic.

chessai: Pushed some docs for this, let me know what you think about the wording/if anything is missing

Control/Monad/Writer/Class.hs
```diff
@@ -204,3 +208,26 @@ instance MonadWriter w m => MonadWriter w (Strict.StateT s m) where
     tell   = lift . tell
     listen = Strict.liftListen listen
     pass   = Strict.liftPass pass
+
+#if MIN_VERSION_transformers(0,5,3)
+-- | There are two valid instances for 'AccumT'. It could either:
+--
+--   1. Lifts the operations to the inner @MonadWriter@
```
chessai, line 215:
```suggestion
--   1. Lift the operations to the inner @MonadWriter@
```

typo

chessai:
I would welcome more reviews, to be safe. Anyone?

chessai:
Okay, now on to #81

PR #82, opened 2020-10-11
Add modifyError  :: MonadError e' m => (e -> e') -> ExceptT e m a -> m a

akrmn:
I recently came to this as a handy way to adapt a function from one `MonadError` to another. I couldn't find anything with this signature on hoogle, so I thought I'd add it here. I'm not 100% sure about the name, so any suggestions are welcome. I hope the docs are clear enough.

### DIFF
--- a/Control/Monad/Error/Class.hs
+++ b/Control/Monad/Error/Class.hs
@@ -49,10 +49,11 @@ module Control.Monad.Error.Class (
     withError,
     handleError,
     mapError,
+    modifyError,
   ) where
 
 import Control.Monad.Trans.Except (ExceptT)
-import qualified Control.Monad.Trans.Except as ExceptT (throwE, catchE)
+import qualified Control.Monad.Trans.Except as ExceptT (catchE, runExceptT, throwE)
 import Control.Monad.Trans.Identity (IdentityT)
 import qualified Control.Monad.Trans.Identity as Identity
 import Control.Monad.Trans.Maybe (MaybeT)
@@ -211,7 +212,7 @@ tryError action = (Right <$> action) `catchError` (pure . Left)
 -- | 'MonadError' analogue to the 'withExceptT' function.
 -- Modify the value (but not the type) of an error.  The type is
 -- fixed because of the functional dependency @m -> e@.  If you need
--- to change the type of @e@ use 'mapError'.
+-- to change the type of @e@ use 'mapError' or 'modifyError'.
 withError :: MonadError e m => (e -> e) -> m a -> m a
 withError f action = tryError action >>= either (throwError . f) pure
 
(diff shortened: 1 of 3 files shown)


akrmn:
hi @ekmett @hvr I'd really appreciate getting some feedback on this, who would be the right person for that? thanks!

ekmett:
A couple of thoughts:

I have no real objection to adding something with this functionality.

However, the name doesn’t quite follow mtl/transformer conventions where lifts are used in a stylized manner to raise operations from lower monads in the stack (and are supplied by transformers rather than mtl).

akrmn:
> I have no real objection to adding something with this functionality.

thanks!

> However, the name doesn’t quite follow mtl/transformer conventions where lifts are used in a stylized manner to raise operations from lower monads in the stack (and are supplied by transformers rather than mtl).

I think this would have to be defined in `mtl`, though, since the type has to mention `MonadError`, which is defined here. Regarding the name, I absolutely agree, naming things is hard indeed. Perhaps

- `exceptToError` by analogy with `exceptToMaybeT` (though the `Error` suffix doesn't quite convey that the return value is `m a` for any `MonadError _ m`)?

- `exceptToMonadError` like above, but makes a bit clearer what it does (though I don't think I've ever seen the word `Monad` in a function name)

- `withError'` by analogy with `withError`, the `'` indicating that the error type can change?

- `rethrow` could also work, there's no analogy with existing functions but the name is memorable and perhaps what users would try to look for. 

- `rethrowExceptT` like above, but indicating the input type

emilypi:
@akrmn I like this functionality. Can I suggest `modifyError` or `translateError` (or some variation thereof)? Then  you have the intuition `withError = modifyError id`.

chessai:
Can this be renamed to `modifyError`? Naming is really the only thing keeping it unmerged.

akrmn:
@emilypi @chessai thanks for the suggestion, I'll rename it to `modifyError`.

emilypi:
Needs a resolution and a changelog entry and we're good

emilypi (approved):

akrmn:
hey @emilypi thanks for taking this forward

Control/Monad/Error/Class.hs
```diff
@@ -225,3 +226,42 @@ handleError = flip catchError
...
+Since the error types do match, @performDatabaseQuery'@ _can_ be used in @app@,
+assuming all other constraints carry over.
+
+This works by instantiating the @m@ in the type of @performDatabaseQuery@ to
+@ExceptT DatabaseError m'@, which satisfies the @MonadError DatabaseError@
+constraint. Immediately, the @ExceptT DatabaseError@ layer is unwrapped,
+producing 'Either' a @DatabaseError@ or a @PersistedValue@. If it's the former,
+the error is wrapped in @MkDatabaseError@ and re-thrown in the inner monad,
+otherwise the result value is returned.
+-}
```
kozross, line 265: Needs an `@since`.

kozross (changes requested):

Control/Monad/Error/Class.hs
emilypi: Very agreed.

akrmn: I can add it; since it's a non-breaking change I guess it should it be `2.3.1`, is that correct?

kozross: @akrmn Yes.

akrmn: Done, I also rebased the changes since I didn't understand the merge conflicts I was getting

akrmn:
I've just added the changelog entry, sorry I forgot to do it earlier

kozross (approved):

emilypi:
Thanks for bearing with us @akrmn and thanks for the PR :)

Issue #86, opened 2021-03-07
Mtl 2.3 Release Prep 


emilypi:
This is release prep for the 2.3 release of `mtl`. 

## TODO 
---

- [x] Add existing typeclass instances for `SelectT`
- [x] `MonadAccum` (including documentation)
- [x] `MonadSelect` (including documentation)
- [x] Compatibility with transformers-0.6
- [ ] Changelog entries 
- [x] Remove all re-exports
- [x] Version narrowing
- [ ] Cut a release

turion:
I can send a `MonadAccum` PR if people would want this.

chessai:
Edited the list to reflect a few new things that need to be done

chessai:
Confirmed that mtl master can build with transformers-0.6

jberryman:
Would someone be able to cut a release with what's already in `master`? It's a shame basic stuff like #67 has been sitting unreleased for years.

emilypi:
@jberryman we may be able to yeah. I think now that the CLC elections are done and we're resolving procedural issues, the next steps are for @chessai and I to sync on when we have time to address this. I'm leaning towards just cutting a release currently. Stay tuned.

turion:
Sorry, I'm behind with polishing `MonadAccum`, but I hope it's not holding up the release? I believe it's not necessary to ship it with the next version. But in case it is, I can speed up.

phadej:
friendly fortnightly ping

matobet:
Could we, in theory, cut a 2.3.0 release with the `transformers` bound bump to let the ecosystem upgrade ASAP and deliver the rest of the checklist items in subsequent 2.3.z versions?

chessai:
This is already being done. I cut rc1 two days ago, and rc2 is coming
tomorrow.

Thanks everyone for the continued push!

phadej:
@chessai that release doesn't seem to allow `transformers-0.6`

chessai:
> @chessai that release candidate doesn't seem to allow `transformers-0.6`

Yeah, that's what rc2 is for. I forgot to bump the bounds before cutting rc1 (though I did already test building with transformers-0.6).

phadej:
@chessai what a reason to not do https://github.com/haskell/mtl/issues/68 in `mtl-2.3`. I'm afraid that `mtl-3` will never happen.

mrkkrp:
Is there any estimate when `mtl-2.3` will be released?

chessai:
[...]

PR #89, opened 2021-07-16
Add MonadAccum

turion:
See #86

### DIFF
--- a/Control/Monad/Accum.hs
+++ b/Control/Monad/Accum.hs
@@ -1,0 +1,30 @@
+{-# LANGUAGE CPP #-}
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  Control.Monad.Accum
+-- Copyright   :  (c) Manuel Bärenz,
+-- License     :  BSD-style (see the file LICENSE)
+--
+-- Maintainer  :  libraries@haskell.org
+-- Stability   :  experimental
+-- Portability :  non-portable (multi-param classes, functional dependencies)
+--
+-- The MonadAccum class.
+--
+--      Inspired by the paper
+--      /Functional Programming with Overloading and Higher-Order Polymorphism/,
+--        Mark P Jones (<http://web.cecs.pdx.edu/~mpj/pubs/springschool.html>)
+--          Advanced School of Functional Programming, 1995.
+-----------------------------------------------------------------------------
+
+module Control.Monad.Accum (
+    module Control.Monad.Accum.Class,
+#if MIN_VERSION_transformers(0,5,3)
+    module Control.Monad.Trans.Accum,
+#endif
+  ) where
+
+#if MIN_VERSION_transformers(0,5,3)
+import Control.Monad.Trans.Accum (AccumT (..), runAccum, runAccumT, execAccum, execAccumT, evalAccum, evalAccumT, mapAccum, mapAccumT)
+#endif
+import Control.Monad.Accum.Class
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -7,6 +7,7 @@ Unreleased
 * Remove `Control.Monad.List` and `Control.Monad.Error`
 * Remove instances of deprecated `ListT` and `ErrorT`
 * Add instances for `Control.Monad.Trans.Accum` and `Control.Monad.Trans.Select`
+* Add `MonadAccum` and corresponding instances
 
 2.2.2
 -----
--- a/Control/Monad/Accum/Class.hs
+++ b/Control/Monad/Accum/Class.hs
@@ -1,0 +1,174 @@
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE FunctionalDependencies #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE TupleSections #-}
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  Control.Monad.Accum.Class
+-- Copyright   :  (c) Manuel Bärenz,
+-- License     :  BSD-style (see the file LICENSE)
+--
+-- Maintainer  :  libraries@haskell.org
+-- Stability   :  experimental
+-- Portability :  non-portable (multi-param classes, functional dependencies)
+--
+-- The MonadAccum class.
+--
+--      Inspired by the paper
+--      /Functional Programming with Overloading and Higher-Order Polymorphism/,
+--        Mark P Jones (<http://web.cecs.pdx.edu/~mpj/pubs/springschool.html>)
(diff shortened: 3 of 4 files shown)


chessai:
This looks good, but it needs a changelog entry

turion:
Weird, CI is not triggered...

chessai:
> Weird, CI is not triggered...

I think our travis credits ran out. I am preparing a PR to switch to GitHub actions.

chessai:
Can you rebase on top of master to get CI changes?

Control/Monad/Accum.hs
```diff
@@ -0,0 +1,23 @@
...
+-- The MonadAccum class.
+--
+--      Inspired by the paper
+--      /Functional Programming with Overloading and Higher-Order Polymorphism/,
+--        Mark P Jones (<http://web.cecs.pdx.edu/~mpj/pubs/springschool.html>)
+--          Advanced School of Functional Programming, 1995.
+-----------------------------------------------------------------------------
+
+module Control.Monad.Accum (
+    module Control.Monad.Accum.Class
```
chessai, line 20: AccumT should be exported here, along with runAccumT, execAccumT, evalAccumT, and mapAccumT

turion:
> Can you rebase on top of master to get CI changes?

Done :) I think it's waiting for you to trigger the CI.

turion:
I'm expecting some older versions to fail, but I think it's easier to test that on CI than to install all the old versions of GHC.

chessai:
I have triggered CI.

turion:
Getting the CPP right is really painful. Is there a good trick to run the whole CI locally somehow? Possibly using nix?

chessai:
Pinging @emilypi @gwils @ekmett for review.

Control/Monad/Accum.hs
```diff
@@ -0,0 +1,30 @@
...
+--
+-- The MonadAccum class.
+--
+--      Inspired by the paper
+--      /Functional Programming with Overloading and Higher-Order Polymorphism/,
+--        Mark P Jones (<http://web.cecs.pdx.edu/~mpj/pubs/springschool.html>)
+--          Advanced School of Functional Programming, 1995.
+-----------------------------------------------------------------------------
+
+module Control.Monad.Accum (
```
chessai, line 20: Instead of re-exporting the module, we need the haddocks to look like others. In order to do that, we need three sections:
* MonadAccum class
* The Accum monad
* The AccumT monad transformer

Please refer to eg the documentation for Control.Monad.Reader

gwils:
Does `MonadAccum` have laws? If so, I would like to see them documented. Do we know whether these instances obey them?
I skimmed the paper linked in `Control.Monad.Accum`, and it looks it does not mention acccumulation monads. Why is it linked?

turion:
> Does MonadAccum have laws? If so, I would like to see them documented.

I thought about some laws, also see this discussion if you're interested: https://github.com/fused-effects/fused-effects/pull/391#discussion_r642113261

It's not clear to me how to generalise them optimally to `mtl`. Possibly we can phrase them by how `add` and `look` interact:

## `add` changes the environment for `look`
```haskell
lhs == rhs
  where
    lhs = add w >> look
    rhs = do
      w' <- look
      add w
      return $ w' <> w
```
## `look`ing twice yields the same
```haskell
(==) <$> look <*> look == return true
```

## `add`ing is monoidal
```haskell
add w >> add w' == add (w <> w')
add mempty == return
```

> Do we know whether these instances obey them?

For `AccumT` this is fairly simple to check (if I haven't done a typo in the above). 

> I skimmed the paper linked in Control.Monad.Accum, and it looks it does not mention acccumulation monads. Why is it linked?

That paper introduces transformers in general, not this specific one which was discovered later.

gwils:
Thank you for putting this patch together. I don't have any problem with the code, but I'm not really sold on adding `MonadAccum` to `mtl`. Maybe @chessai can help me.

Where did `Accum` come from? Does we know of anyone using it for anything? It looks like an awkward point in the design space.  Does this have some helpful special property that I've missed?
I would prefer to use an [update monad](https://www.schoolofhaskell.com/user/edwardk/heap-of-successes) in many cases instead of this, or otherwise use `State` or `Writer` directly.

turion:
> Where did Accum come from?

`AccumT` was introduced in `transformers` here: https://hub.darcs.net/ross/transformers/issue/24 It was known _and_ used before that.

> Does we know of anyone using it for anything?

I've used it myself in a few places in the past, even in production. Some reasons why it's not so common as other transformers might be:

* It arrived later to the party, so all old tutorials, teaching materials and libraries don't refer to it.
* It is not supported by `mtl` (and other effect libraries like `fused-effects`) yet, so using it is currently a maintability risk.

The second one is kind of a self-fulfilling prophecy. `mtl` doesn't support `Accum`, so noone uses it, so noone tries to add it. In fact I came to create this PR precisely because I used it in my code, and then at some point couldn't work with my transformer stack anymore because `AccumT` was poorly supported, both in lacking instances and the `MonadAccum` class. (@chessai has improved the situation with https://github.com/haskell/mtl/pull/80 since.)

> It looks like an awkward point in the design space. Does this have some helpful special property that I've missed? I would prefer to use an update monad in many cases instead of this, or otherwise use State or Writer directly.

`AccumT` sits in between `WriterT` and `StateT`, and has advantages and disadvantages over both.

Monad | `AccumT` Advantage | `AccumT` Disadvantage
---------- | -------------- | -------------------
`WriterT` | Can read the log (as a monadic action) | `WriterT` is functorial in the log type, `AccumT` not
`StateT` | Extra guarantee that old logs are not discarded | `StateT` is more flexible

Typical use cases for me are counters, or handing out unique keys/ids/tokens.

I wasn't familiar with update monads, so not sure how those relate. To me it seems that `type AccumT w m a = UpdateT w w m a`, where the monoid acts on the state by multiplication. So maybe `AccumT` is a special case of `UpdateT`. Nevertheless I believe that `UpdateT` doesn't even exist in any popular library. But `AccumT` is part of one of the most popular and used Haskell libraries.

Should `mtl` support it? I believe yes. The github readme says:

> MTL is a collection of monad classes, extending the transformers package

So if `mtl` extends `transformers`, it should ideally have a class for every transformer.

Control/Monad/Accum.hs
```diff
@@ -0,0 +1,30 @@
...
+-- Copyright   :  (c) Manuel Bärenz,
+-- License     :  BSD-style (see the file LICENSE)
+--
+-- Maintainer  :  libraries@haskell.org
+-- Stability   :  experimental
+-- Portability :  non-portable (multi-param classes, functional dependencies)
+--
+-- The MonadAccum class.
+--
+--      Inspired by the paper
```
gwils, line 14: I think you should remove this section  given this transformer is not inspired by that paper. Some similar modules don't have this disclaimer in them, such as `Control.Monad.Identity`.

kozross:
Superceded by #109

PR #97, opened 2021-11-13
Switch back to haskell-ci

phadej:
Add GHC-7 jobs, add base lower bound


chessai:
What is the justification for haskell-ci? The ci.yml becomes much more verbose.

Additionally, do you think we should move back to supporting 7.x?

phadej:
> Additionally, do you think we should move back to supporting 7.x?

There is no reason to not to. They just work.

>  The ci.yml becomes much more verbose.

You need to care only about the first lines:

```
# This GitHub workflow config has been generated by a script via
#
#   haskell-ci 'github' '--installed=+all -transformers' '--haddock-jobs=>=8.0' 'mtl.cabal'
#
# To regenerate the script (for example after adjusting tested-with) run
#
#   haskell-ci regenerate
#
# For more information, see https://github.com/haskell-CI/haskell-ci
```

chessai:
Thanks!

PR #106, opened 2022-04-06
Add note to Haddock about possible removals of re-exports

tomjaguarpaw:
If one long term possibility is that we remove certain re-exports from `mtl` modules it seems like the documentation should warn about this.  Of course, this is not as good as deprecating the re-exports, but that functionality doesn't exist, and in its absence it seems like this would be an important way to communicate future potential breakage to users.

screenshot

### DIFF
--- a/Control/Monad/Cont.hs
+++ b/Control/Monad/Cont.hs
@@ -64,8 +64,6 @@ module Control.Monad.Cont (
     evalContT,
     mapContT,
     withContT,
-    module Control.Monad,
-    module Control.Monad.Trans,
     -- * Example 1: Simple Continuation Usage
     -- $simpleContExample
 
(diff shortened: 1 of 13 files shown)


kozross:
@tomjaguarpaw: Given [this](https://github.com/haskell/core-libraries-committee/issues/52#issuecomment-1094103712) and [this](https://github.com/haskell/mtl/issues/86#issue-823982952), I believe we should go further, tear off the band-aid, and just remove the re-exports completely.

tomjaguarpaw:
I'm inclined to agree that we should completely remove the re-exports in the next major release, but can we make a point release first (or indeed "at the same time") containing all the non-breaking changes that it can? That would provide a smoother upgrade path.

kozross:
I think we should just take the big step now. Splitting it up like this is unlikely to be much better, and I honestly can't see good reason for it. Improvement is long overdue, and this is a _very_ bad legacy decision.

tomjaguarpaw:
As you wish. I am in support of [Bodigrim's comments](https://github.com/haskell/mtl/issues/101#issuecomment-1005139255) on #101. Personally I think it it would be preferable for the band aid to be ripped off quickly with respect to the passage of time but slowly with respect to the increase of version numbers. As Bodigrim says, 2.3 can be released at the same time as 2.4. I believe that would provide a smoother experience for users. But ultimately the choice rests with you.

emilypi (approved):
Pending merge with master

kozross:
@emilypi I think this has been superceded already?

emilypi:
Yeah - you did write the thing that's true.

PR #108, opened 2022-05-01
Re-remove re-exports, minimal bounds, CI

kozross:
This completes multiple required milestones in #86:

* Re-exports of `Control.Monad`, `Control.Monad.Fix` and `Data.Monoid` are gone.
* We now require at least GHC 8.6 and `cabal-install` 3.0.
* New Github Actions-based CI holding this in place.
* All CPP is gone.
* Require at least `transformers-0.5.6`.

@emilypi I've asked about whether we dealt with other re-exports or not: I assume you only meant those three in your item, but I had to check. This is a necessary preliminary to other work.


emilypi (approved):
herculean 💪 

Re: 

> I've asked about whether we dealt with other re-exports or not: I assume you only meant those three in your item, but I had to check. This is a necessary preliminary to other work.

Deferring to @chessai, but I believe those three were the offending re-exports.

mtl.cabal
```diff
@@ -59,24 +52,15 @@ Library
...
-    CPP
-    MultiParamTypeClasses
-    FunctionalDependencies
-    FlexibleInstances
-    UndecidableInstances
-
-  -- This is a SafeHaskell safeguard (pun intended) to explicitly declare the API contract of `mtl`
-  -- GHC versions before 7.4 were hopelessly broken or incapable of SafeHaskell
-  if impl(ghc >= 7.4)
-    default-extensions: Safe
```
phadej, line 75: Removal of `default-extensions: Safe` may accidentally make some modules unsafe. Please add it back.

PR #109, opened 2022-05-02
MonadAccum

kozross:
This defines a type class, along with a helper `DerivingVia` type, a bunch of instances, documentation and laws, to address the corresponding entry in #86.

This is WIP for now, as I want to add some details about the meaning of the various instances, which may not be 100% obvious to a first-time reader.

### DIFF
--- a/Control/Monad/Accum.hs
+++ b/Control/Monad/Accum.hs
@@ -1,0 +1,83 @@
+{-# LANGUAGE FunctionalDependencies #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TupleSections #-}
+
+-- | Module: Control.Monad.Accum
+-- Copyright: (C) Koz Ross 2022
+-- License: BSD-3-Clause (see the LICENSE file)
+-- Maintainer: koz.ross@retro-freedom.nz
+-- Stability: Experimental
+-- Portability: GHC only
+--
+-- [Computation type:] Accumulation (either append-only state, or writer with
+-- the ability to read all previous input).
+--
+-- [Binding strategy:] Binding a function to a monadic value monoidally
+-- accumulates the subcomputations (that is, using '<>').
+--
+-- [Useful for:] Logging, patch-style tracking.
+--
+-- [Zero and plus:] None.
+--
+-- [Example type:] @'Accum' w a@
+module Control.Monad.Accum
+  ( MonadAccum (..),
+    looks,
+  )
+where
+
+import Data.Kind (Type)
+
+-- | The capability to accumulate. This can be seen in one of two ways:
+--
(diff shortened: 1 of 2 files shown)


Control/Monad/Accum.hs
```diff
@@ -0,0 +1,247 @@
...
+-- the ability to read all previous input).
+--
+-- [Binding strategy:] Binding a function to a monadic value monoidally
+-- accumulates the subcomputations (that is, using '<>').
+--
+-- [Useful for:] Logging, patch-style tracking.
+--
+-- [Zero and plus:] None.
+--
+-- [Example type:] @'Accum' w a@
```
emilypi, line 27: I actually kind of like this doc style. 👍

emilypi (commented):
WIP, needs some expanded docs, but otherwise LGTM

Control/Monad/Accum.hs
```diff
@@ -0,0 +1,247 @@
...
+-- all previous 'tell's.
+--
+-- = Laws
+--
+-- 'accum' should obey the following:
+--
+-- 1. @'accum' ('const' (x, 'mempty'))@ @=@ @'pure' x@
+-- 2. @'accum' f '*>' 'accum' g@ @=@
+-- @'accum' '$' \acc -> let (_, v) = f acc
+--                          (res, w) = g (acc '<>' v) in (res, v '<>' w)@
```
emilypi, line 74: I would name these laws: 

1.  `[Pure-Quot]` (in the sense of `pure x` being quotiented with `(x, mempty)`)
2. `[Const-Accum]` because it is what it says it is.

Control/Monad/Accum.hs
kozross: Can't claim any credit here. I'm just using what already existed for (at least) `MonadCont`.

Control/Monad/Accum.hs
kozross: We'd need to name all the others too, then. Good idea though.

turion:
I guess you're aware of https://github.com/haskell/mtl/pull/89, on which I took a break because I didn't quite get how the docs were supposed to be structured.

turion:
And, I should add, I can finish #89 if that saves you work & brain capacity.

kozross:
@turion At this point I'm nearly done anyway - I just have some docs to add and that's it. But thanks for offering!

emilypi:
@kozross we should add @turion as a coauthor on the squash/rebased commit to master just to acknowledge their additions. We (chessai and myself) let their PR languish, and that's why it didn't go in.

kozross:
@emilypi My work is clean-room, though I'm happy to add @turion into the list of authors.

kozross:
@emilypi and @chessai - I wrote the docs I wanted, and addressed all the things needing addressing. I couldn't do a proper description of what the `SelectT` stack does, but that's because I don't even know what `SelectT` does in general, as @emilypi is aware.

PR #112, opened 2022-05-06
MonadSelect, laws, docs, helpers

kozross:
This defines a `MonadSelect` and `LiftingSelect`, similar to `MonadAccum` and `LiftingAccum`. I'm not 100% confident on laws and effect descriptions, but this is the best I could come up with.

Completes all outstanding tasks in #86 to the best of my understanding.

### DIFF
--- a/Control/Monad/Select.hs
+++ b/Control/Monad/Select.hs
@@ -1,0 +1,316 @@
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE FunctionalDependencies #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE Trustworthy #-}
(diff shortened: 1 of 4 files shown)


Control/Monad/Select.hs
```diff
@@ -0,0 +1,316 @@
...
+
+-- | Module: Control.Monad.Select
+-- Copyright: (C) Koz Ross 2022
+-- License: BSD-3-Clause (see the LICENSE file)
+-- Maintainer: koz.ross@retro-freedom.nz
+-- Stability: Experimental
+-- Portability: GHC only
+--
+-- [Computation type:] Backtracking search, with @r@ as a \'ranking\' or
+-- \'evaluation\' type.
```
emilypi, line 20: `r` for "ranking" is actually quite a nice way of thinking about it.

emilypi (approved):

Control/Monad/Select.hs
kozross: It's buried 20 pages into the Escardo et al paper.

PR #121, opened 2022-05-09
Restore ExceptT re-exports

### LINKED ISSUE #117: `mtl-2.3` dropped `ExceptT`
Apparently, `mtl-2.3` dropped `ExceptT`, see e.g.:
- https://github.com/haskell/happy/commit/cc0fc789a19b9c3bdb1b1ed29bf910f386cefdf7

This is a bit surprising, as e.g. it has [`ReaderT`](https://hackage.haskell.org/package/mtl-2.3/docs/Control-Monad-Reader.html#t:ReaderT) and more of that kind: `WriterT`, `RWST`, `IdentityT` ...

Note that this drop is inconsistent with the documentation of `mtl-2.3`.  E.g. the following example now produces a scope error:
- https://hackage.haskell.org/package/mtl-2.3/docs/Control-Monad-Except.html#g:3

The changelog also does not mention this:
- https://hackage.haskell.org/package/mtl-2.3/changelog

So maybe `ExceptT` was dropped on error?

(Let me add that I am surprised that this release came without a RC...)

kozross:
Closes #117.

### DIFF
--- a/Control/Monad/Except.hs
+++ b/Control/Monad/Except.hs
@@ -42,6 +42,15 @@ module Control.Monad.Except
     Error.withError,
     Error.handleError,
     Error.mapError,
+    -- * The ExceptT monad transformer
+    Except.ExceptT(ExceptT),
+    Except.Except,
+    Except.runExceptT,
+    Except.mapExceptT,
+    Except.withExceptT,
+    Except.runExcept,
+    Except.mapExcept,
+    Except.withExcept,
     -- * Example 1: Custom Error Data Type
     -- $customErrorExample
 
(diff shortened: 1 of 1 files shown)


emilypi (approved):

Control/Monad/Except.hs
```diff
@@ -38,6 +38,17 @@ module Control.Monad.Except
     -- * Monads with error handling
     Error.MonadError(..),
     Error.liftEither,
+    -- * The ExceptT monad transformer
```
andreasabel, line 41: How does this render in haddock?  It seems that `tryError` and the like would belong to `MonadError` rather than `Except`.
Confer this with what it was previously: https://github.com/haskell/mtl/blame/ecb525a5ba3521d0059eb2396a376b2e91557cff/Control/Monad/Except.hs#L39

I think the easiest path would be to restore the order to what it was before `ExceptT` was removed.

andreasabel (changes requested):

Control/Monad/Except.hs
kozross: This has now been done.

andreasabel (approved):
Thanks!

Issue #120, opened 2022-05-08
Redundant constraint Monoid w in GHC 9.4 build


Bodigrim:
```
$ cabal build -w ghc-9.4.0.20220501
...
[15 of 24] Compiling Control.Monad.Writer.Class

Control/Monad/Writer/Class.hs:201:3: warning: [-Wredundant-constraints]
    • Redundant constraint: Monoid w
    • In the instance declaration for ‘MonadWriter w (AccumT w m)’
    |
201 |   ( Monoid w
    |   ^^^^^^^^^^...
```

kozross:
This appears to be a typo in any case: there's no reason why the type of the writer effect has to match that of the accumulation effect.

PR #122, opened 2022-05-10
Separate type of writer and accumulator

kozross:
Previously, the instance required that both the `MonadWriter` and `AccumT` had the same type `w` as their accumulated value, even though in theory, they could be different. This was prompted by #120.

I don't have (easy) access to GHC 9.4. @Bodigrim - do you still get the redundant constraint warning? @emilypi - we should merge this in any case, as the original instance is definitely not what we intended.

### DIFF
--- a/Control/Monad/Writer/Class.hs
+++ b/Control/Monad/Writer/Class.hs
@@ -198,9 +198,9 @@ instance MonadWriter w m => MonadWriter w (Strict.StateT s m) where
 --
 -- @since 2.3
 instance
-  ( Monoid w
+  ( Monoid w'
   , MonadWriter w m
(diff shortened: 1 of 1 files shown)


emilypi (approved):

Control/Monad/Writer/Class.hs
```diff
@@ -199,8 +199,9 @@ instance MonadWriter w m => MonadWriter w (Strict.StateT s m) where
 -- @since 2.3
 instance
   ( Monoid w
```
Bodigrim, line 201: You can remove this line now.

kozross: This is now done.

Bodigrim (approved):

PR #125, opened 2022-05-21
MonadAccum law tests

kozross:
This relates to [this comment here](https://github.com/haskell/mtl/issues/117#issuecomment-1120504674). Specifically, these tests use property-based testing to exercise all the stated laws of `MonadAccum`, with a large enough test count to be meaningful.


andreasabel (commented):
How can I help here?

Kudos for the tests!

I don't know `MonadAccum`, so I might not be the best person to review this.

kozross:
@andreasabel I mostly tagged you as a reviewer because you suggested/requested some tests. Was this what you had in mind?

I started with `MonadAccum` for two reasons: it's one of the (currently only two) type classes provided by `mtl` which even states any laws, and of the two, it is by far the easiest, both in terms of the laws themselves, and also to test.

andreasabel:
Well done!

> I started with `MonadAccum` for two reasons: it's one of the (currently only two) type classes provided by `mtl` which even states any laws, and of the two, it is by far the easiest, both in terms of the laws themselves, and also to test.

Testing the laws is of course the gold standard!

Even simple unit tests that cover the functionality are useful, detecting basic problems like missing exports, unintended loops, etc.

emilypi (approved):

PR #132, opened 2022-08-31
fix tag in mtl-2.3.x migration guide

brprice:
Since mtl-2.3.1 has not yet been tagged, I cannot be sure that this fix is correct... . However, going by historical precedent (all other tags start with a `v`, including `v2.3.1-rc1`) I suspect that this was a typo.


emilypi:
approved thanks @brprice

PR #141, opened 2023-01-13
`AccumT` is a regular transformer and should have a regular `MonadAccum` instance.

jumper149:
I think this was accidentaly copy-pasted, see #140 .

I don't see any problem with doing it this way.
Maybe you can spot anything?


jumper149:
Moving this here just so that it doesn't get lost (from #140):

> 
> > > Edit: I don't understand why the MonadAccum instance for AccumT also uses Identity as a base monad. I think it should be fine with any monad, right?
> > 
> > 
> > Not unless you want to overlap on the monoid instance. Every Monad may admit any number of accumulating monoids, so the general instance is not supplied. This is a case where we ask not "can we?" but "_should_ we?".
> 
> This discussion should move to #141.

---

My motivation was `WriterT` in this case.

```haskell
instance (Monoid w, Monad m) => MonadWriter (WriterT w m)
```

I also feel like this is the way, that all transformers instances work (except for `SelectT`).
Even `ContT` is sort of like that.

---

So if you would argue to keep `Monoid w => MonadAccum w (AccumT w Identity)`, then you should also argue for `Monoid w => MonadWriter (WriterT w Identity)`, right?

Well I think the way that `WriterT` does it currently is better.

emilypi:
> My motivation was WriterT in this case.
> I also feel like this is the way, that all transformers instances work (except for SelectT).
Even ContT is sort of like that.

Sorry, I'm confused as to why you feel this is "the way"? What way? Are you making the argument that it's always the case that `(Monoid w, Monad m) => ` work as constraints for `Writer` and `AccumT`?  Have you checked the laws? I'm finding that `(->) t` is a counter example to the laws for `MonadAccum`.  I suspect that this is the case for `Writer` as well, but I'll have to write the equations down.

jumper149:
> Have you checked the laws?

I did the work: https://felixspringer.xyz/homepage/blog/accumtsMonadaccumInstance

---

> I'm finding that `(->) t` is a counter example to the laws for `MonadAccum` for all monoids except `()`.

I didn't check this in particular, but considering that I just proved it for any base monad your finding cannot be true (unless I made a mistake of course, so please thoroughly check it).

Lysxia:
This looks like a fine change to me. I'm also puzzled why `Identity` is in this instance in the first place.

emilypi:
@jumper149 I believe i was looking at the PR last night and proved something else for something entirely unrelated to the PR. I checked again and yes, your proof is fine, thanks for providing it. It seems like this would be fine!

turion (approved):

Futarimiti:
Any updates on this? I wonder why is this not merged still after 2 years

turion:
I believe this could be merged as-is, right?

PR #143, opened 2023-02-20
Add @since annotations for tryError,withError,handleError,mapError

bezirg:


kozross (approved):

emilypi (changes requested):
I'd rather not change the comment style of the existing module and just keep with `-- |`. If you could revert the major changes and just add a `@since` entry to the existing comments, it'd makes this PR much smaller.

emilypi (approved):

emilypi:
Thanks @bezirg !

PR #146, opened 2023-12-18
add `overwrite` method to `MonadWriter`

wygulmage:
`overwrite` is equivalent to `mapWriter` without output type change. It allows default definitions of all other methods. It has a default definition in terms of `pass` and `listen`.

Normally I would just make `mapWriter` a method, but unfortunately that would clash with the exports of CPS, Lazy, and Strict. Currently the method order of `overwrite` is sane for the `(,)` instance but swapped compared to `mapWriter`. Would it make more sense to keep the API exactly the same?

Is the benefit worth the API maintenance cost and learning burden?


wygulmage:
I just noticed that this repo is completely inactive; closing pull request.

PR #149, opened 2024-04-17
Bypass Coercible Type for Deriving Monad* Classes of New Transformers Composed of Transformers

BebeSparkelSparkel:
When creating a new transformer it is often composed of other transformers. The transformers that it is composed of usually have specific purposes that are unrelated to the more general uses of the Monad* classes and typically you still want to define the Monad* instances for the underlying monad.

My real world use case is the

```haskell
newtype FileT w m a = FileT (ReaderT Handle m a)
```

where I still want to have MonadReader but not for reading Handle but rather the r for m `instance MonadReader r m => MonadReader r (FileT w m)`

This is a bit annoying for this case but even more so for transformers wrapping RWS. In just the ReaderT  case the following must be defined

```haskell
instance MonadReader r m => MonadReader r (FileT w m) where
  local f (FileT (ReaderT g)) = FileT . ReaderT $ local f . g 
  reader = lift . reader
```

DerivingVia seems to be a better option for this by using `Bypass`

```haskell
deriving via ByPass (ReaderT Handle) m instance MonadReader r m => MonadReader r (FileT w m)
```

There could probably a more succinct way to write the derivation but this is what I have so far.

### DIFF
--- a/Control/Monad/Bypass.hs
+++ b/Control/Monad/Bypass.hs
@@ -1,0 +1,40 @@
+{-# LANGUAGE StandaloneKindSignatures #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE DerivingVia #-}
+module Control.Monad.Bypass
+  ( Bypass
+  ) where
+
+import Control.Monad.Reader
+import Control.Monad.State
+import Control.Monad.Writer
+import Data.Kind (Type)
+
+type Bypass :: ((Type -> Type) -> Type -> Type) -> (Type -> Type) -> Type -> Type
+newtype Bypass t m a = Bypass (t m a) deriving (Functor, Applicative, Monad, MonadTrans)
+
+instance MonadReader r m => MonadReader r (Bypass (ReaderT r') m) where
+  ask = lift ask
+  local f (Bypass (ReaderT x)) = Bypass . ReaderT $ local f . x
+  reader = lift . reader
+
+instance (MonadWriter w m, Monoid w') => MonadWriter w (Bypass (WriterT w') m) where
+  writer = lift . writer
+  tell = lift . tell
+  listen (Bypass (WriterT x)) = Bypass $ WriterT $ (\((a, w'), w) -> ((a, w), w')) <$> listen x
+  pass (Bypass (WriterT x)) = Bypass $ WriterT $ (\((a,f),w') -> pass $ return ((a,w'),f)) =<< x
+
+instance MonadState s m => MonadState s (Bypass (StateT s') m) where
+  get = lift get
+  put = lift . put
+  state = lift . state
+
+newtype ExampleT m a = ExampleT (ReaderT Int m a) deriving (Functor, Applicative, Monad)
+
+deriving via Bypass (ReaderT Int) m instance MonadReader r m => MonadReader r (ExampleT m)
+
--- a/mtl.cabal
+++ b/mtl.cabal
@@ -32,6 +32,7 @@ Library
   exposed-modules:
     Control.Monad.Cont
     Control.Monad.Cont.Class
+    Control.Monad.Bypass
     Control.Monad.Error.Class
     Control.Monad.Except
     Control.Monad.Identity


rhendric:
Is the example an off-label use of `MonadReader`? Are the `MonadFoo` classes intended to be generic interfaces for anything that can plausibly represent the concept of fooing, or are they specific tools for accessing the `Control.Monad.Foo.FooT` layer in an `mtl`-style transformer stack, regardless of that layer's depth in the stack? The documentation doesn't seem to make this point one way or the other, but I've always assumed the latter, and this proposal only makes sense in the context of the former.

rhendric:
(Earlier message was a mistake; sorry for noise.)

Is this basically the same concept as `LiftingAccum` and `LiftingSelect`? If so, why not follow that convention, or propose aliasing those newtypes to `Bypass`?

BebeSparkelSparkel:
I appreciate the established convention you pointed out. So, for the case above propose to create `LiftingReader`?

kozross:
The reason we don't have `LiftingX` for every `X` at the moment is because `DerivingVia` wasn't in our support window until recently. I have no objection to adding such a type class for every 'capability type class' we support.

BebeSparkelSparkel:
Is there a way for the current Lifting* convention to work with the various Reader, Writer, State, and RWS? I think it would be annoying to have to differentiate between the various types for the same property.

For example with MonadWriter there are lazy, strict, and cps versions of Writer and RWS. Could there be just one LiftingWriter for all of them? If so, which module would it be defined in?

BebeSparkelSparkel:
I have implemented a partial example for LiftingReader. Can you check if this is what you are looking for?

Control/Monad/Reader/Class.hs
```diff
@@ -202,3 +206,17 @@ instance
...
+
+instance MonadReader r m => MonadReader r (LiftingReader (ReaderT r') m) where
+  ask = lift ask
+  local f (LiftingReader (ReaderT.ReaderT x)) = LiftingReader . ReaderT.ReaderT $ local f . x
+  reader = lift . reader
+
+instance (MonadReader r m, Monoid w) => MonadReader r (LiftingReader (LazyRWS.RWST r' w s) m) where
+  ask = lift ask
+  local f (LiftingReader (LazyRWS.RWST x)) = LiftingReader . LazyRWS.RWST $ \r s -> local f $ x r s
+  reader = lift . reader
```
rhendric, line 222:
```suggestion
instance MonadReader r m => MonadReader r (LiftingReader (ReaderT r') m) where
  ask = lift ask
  local f = mapLiftingReader (ReaderT.mapReaderT (local f))
  reader = lift . reader

instance (MonadReader r m, Monoid w) => MonadReader r (LiftingReader (LazyRWS.RWST r' w s) m) where
  ask = lift ask
  local f = mapLiftingReader (LazyRWS.mapRWST (local f))
  reader = lift . reader

mapLiftingReader :: (t m a -> t n b) -> LiftingReader t m a -> LiftingReader t n b
mapLiftingReader = coerce
```

BebeSparkelSparkel: This github diff view is terrible. Did you only change lines 216 and 221 with mapLiftingReader?

rhendric: Yup.

BebeSparkelSparkel:
Is there anything else that needs to be discussed before I finish this up for MonadReader, MonadWriter, and MonadState?

Control/Monad/State/Class.hs
```diff
@@ -192,3 +196,37 @@ instance MonadState s m => MonadState s (SelectT r m) where
...
+
+instance MonadState s m => MonadState s (LiftingState (Lazy.StateT s') m) where
+  get = lift get
+  put = lift . put
+  state = lift . state
+
+instance MonadState s m => MonadState s (LiftingState (Strict.StateT s') m) where
+  get = lift get
+  put = lift . put
+  state = lift . state
```
rhendric, line 231: These instances could be unified via `MonadTrans`, like the instances for `LiftingAccum` and `LiftingSelect`. (The others unfortunately require more power; I believe `MonadTransControl` from `monad-control` would be sufficient but obviously that's not available here.)

Control/Monad/Writer/Class.hs
```diff
@@ -205,3 +212,80 @@ instance
...
+instance (Monoid w', MonadWriter w m) => MonadWriter w (LiftingWriter (LazyRWS.RWST r w' s) m) where
+  writer = lift . writer
+  tell = lift . tell
+  listen (LiftingWriter (LazyRWS.RWST x)) = LiftingWriter $ LazyRWS.RWST $ \r s -> do
+    ((a, s, w'), w) <- listen $ x r s
+    pure ((a, w), s, w')
+  pass (LiftingWriter (LazyRWS.RWST x)) = LiftingWriter $ LazyRWS.RWST $ \r s -> do
+    (y, s, w') <- x r s
+    a <- pass $ pure y
+    pure (a, s, w')
```
rhendric, line 235: These `pass` implementations are incorrect. Here is a test:

```hs
newtype SneakyWriterT m a = SneakyWriterT { runSneakyWriterT :: Lazy.WriterT String m a }
  deriving (Functor, Applicative, Monad)
  deriving (MonadWriter w) via LiftingWriter (Lazy.WriterT String) m

main :: IO ()
main = print $ runIdentity $ Lazy.runWriterT $ Lazy.runWriterT $ runSneakyWriterT $ pass $ do
  tell "hello"
  pure ((), reverse)
```

This should print `(((),""),"olleh")`, but with your implementation it prints `(((),""),"hello")`.

Avoiding errors like this is one reason why I [suggested](https://github.com/haskell/mtl/pull/149#pullrequestreview-2009506739) using the existing `map*T` functions instead of inlining your own implementations of them. For `MonadWriter`, that would look like:

```hs
sequenceFst :: ((a, b), c) -> ((a, c), b)
sequenceFst ((a, b), c) = ((a, c), b)

sequenceFst23 :: ((a, b, c), d) -> ((a, d), b, c)
sequenceFst23 ((a, b, c), d) = ((a, d), b, c)

sequenceFst32 :: ((a, b), c, d) -> ((a, c, d), b)
sequenceFst32 ((a, b), c, d) = ((a, c, d), b)

mapLiftingWriter :: (t m a -> t m b) -> LiftingWriter t m a -> LiftingWriter t m b
mapLiftingWriter = coerce

instance (Monoid w', MonadWriter w m) => MonadWriter w (LiftingWriter (LazyRWS.RWST r w' s) m) where
  writer = lift . writer
  tell = lift . tell
  listen = mapLiftingWriter (LazyRWS.mapRWST (fmap sequenceFst23 . listen))
  pass = mapLiftingWriter (LazyRWS.mapRWST (pass . fmap sequenceFst32))

-- ... and likewise for the remaining RWST instances

instance (Monoid w', MonadWriter w m) => MonadWriter w (LiftingWriter (Lazy.WriterT w') m) where
  writer = lift . writer
  tell = lift . tell
  listen = mapLiftingWriter (Lazy.mapWriterT (fmap sequenceFst . listen))
  pass = mapLiftingWriter (Lazy.mapWriterT (pass . fmap sequenceFst))

-- ... and likewise for the remaining WriterT instances
```

BebeSparkelSparkel:
I have tried a different approach for `LiftingWriter` and am using a type family and mapping classes. Let me know what you think of this.

rhendric:
I see no glaring issues.

BebeSparkelSparkel:
@kozross Could you check if this is ready to pull?

sjshuck:
I pulled this into my fork.  It needs some superficial cleanup, and I'd probably replace the type families-powered `LiftingWriter` with a newtype for consistency.  But I intend to merge this.

PR #161, opened 2025-09-07
Make it compile with MicroHs

augustss:
Also, CI for mhs if you want that.

### DIFF
--- a/.github/workflows/ci-mhs.yml
+++ b/.github/workflows/ci-mhs.yml
@@ -1,0 +1,48 @@
+name: ci-mhs
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
+
+jobs:
+  build-mhs-mtl:
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: checkout mhs repo
+      uses: actions/checkout@v4
+      with:
+        repository: augustss/MicroHs
+        ref: v0.14.15.0
+        path: mhs
+    - name: make and install mhs
+      run: |
+        cd mhs
+        make minstall
+
+    - name: checkout transformers repo
+      uses: actions/checkout@v4
+      with:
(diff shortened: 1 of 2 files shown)


augustss:
It does support polykinds, but it doesn't infer polykinds. You have to have
a kind signature.

sjshuck:
I wanted to acknowledge @hasufell's bump of this, but I'm not in a hurry to merge anything.  This would be my first merge as mtl maintainer.

My hesitation is due to human considerations, none technical.  I see there's a dependency in the CI on your fork of transformers.  Only having learned about MicroHs this morning, I am envisioning that multiple core libraries will require `#ifdef __MHS__`, and any preexisting CI workflows would need to be extended to include mhs, and then the question of, should those workflows be blocking.  In short, setting a precedent that the core libraries should support mhs going forward.  I'm sure a lot of these discussions are the same for things like GHCJS.

What is the story for expectations of core libraries to support various compilers and versions?  Any links would be appreciated.  Additionally I will bring this up next week in MuniHac to get a feel of the room (literally).  To be clear, I want to merge, but with eyes open.

sjshuck:
Okay I see `__MHS__` macros in the transformers-0.6.2.0 source.  Incidentally I had to download the tarball and extract manually; see https://github.com/haskell/core-libraries-committee/issues/365.

The CI looks good and runs nice and quick.  I'll probably merge as-is.

sjshuck:
@augustss I'd to include mhs in the `tested-with` field in the Cabal file.  Any insights?

augustss:
No. I've not done that

hasufell:
Cabal knows about MicroHs: https://github.com/haskell/cabal/blob/832796bbfd1bf7bf9c47cf78de2a59eef075b9a8/Cabal-syntax/src/Distribution/Compiler.hs#L78-L80

Issue #84, opened 2020-11-26
instance (MonadError e f, MonadError e g) => MonadError e (Product f g)


LightAndLight:
Is there a technical reason why this instance doesn't exist? It seems well-behaved (some sanity-check proofs are at the bottom of the page).

```
fstP :: Product f g a -> f a
fstP (Pair a _) = a

sndP :: Product f g a -> g a
sndP (Pair _ a) = a

instance (MonadError e f, MonadError e g) => MonadError e (Product f g) where
  throwError e = Pair (throwError e) (throwError e)
  catchError (Pair ma mb) f = Pair (catchError ma (fstP . f)) (catchError mb (sndP . f))
```

---

```
catchError m throwError = m

catchError (Pair f g) throwError
= { inline catchError }
Pair (catchError f (fstP . throwError)) (catchError g (sndP . throwError))
= { inline throwError }
Pair 
  (catchError f (fstP . (\e -> Pair (throwError e) (throwError e)))
  (catchError g (sndP . (\e -> Pair (throwError e) (throwError e)))
= { inline (.) }
Pair 
  (catchError f (\e -> fstP (Pair (throwError e) (throwError e)))
  (catchError g (\e -> sndP (Pair (throwError e) (throwError e)))
= { eta-reduction }
Pair 
  (catchError f (\e -> throwError e))
  (catchError g (\e -> throwError e))
= { eta-reduction }
Pair 
  (catchError f throwError)
  (catchError g throwError)
= { hypothesis }
Pair f g
```

```
catchError (throwError e) pure = pure e

catchError (throwError e) pure
= { inline throwError }
catchError (Pair (throwError e) (throwError e)) pure = pure e
= { inline catchError }
Pair (catchError (throwError e) (fstP . pure)) (catchError (throwError e) (sndP . pure))
= { inline pure, (.) }
Pair 
  (catchError (throwError e) (\x -> fstP (Pair (pure x) (pure x)))) 
  (catchError (throwError e) (\x -> sndP (Pair (pure x) (pure x))))
= { eta-reduction }
Pair 
  (catchError (throwError e) (\x -> pure x)) 
  (catchError (throwError e) (\x -> pure x))
= { eta-reduction }
Pair 
  (catchError (throwError e) pure) 
  (catchError (throwError e) pure)
= { hypothesis }
Pair (pure e) (pure e}
= { pure }
pure e
```

sjshuck:
@LightAndLight PR welcome.

[...]

PR #171, opened 2025-10-24
Add instances for functor product

LightAndLight:
Closes https://github.com/haskell/mtl/issues/84

Compatibility: `Data.Functor.Product` was added to `transformers` in version 0.2.1, and to `base` in version `4.9`.

### DIFF
--- a/Control/Monad/Error/Class.hs
+++ b/Control/Monad/Error/Class.hs
@@ -74,6 +74,7 @@ import qualified Control.Monad.Trans.Writer.CPS as CPSWriter
 import Control.Monad.Trans.Class (lift)
 import Control.Exception (IOException, catch, ioError)
 import Control.Monad (Monad ((>>=), (>>)))
+import Data.Functor.Product (Product(..))
 import Data.Monoid (Monoid)
 import Prelude (Either (Left, Right), Maybe (Nothing), either, flip, (.), IO, pure, (<$>))
 
@@ -205,6 +206,13 @@ instance
   ) => MonadError e (AccumT w m) where
     throwError = lift . throwError
     catchError = Accum.liftCatch catchError
+
+instance (MonadError e m, MonadError e n) => MonadError e (Product m n) where
+    throwError e = Pair (throwError e) (throwError e)
+    catchError (Pair ma na) f = Pair (catchError ma (productFst . f)) (catchError na (productSnd . f))
+        where
+            productFst (Pair a _) = a
+            productSnd (Pair _ b) = b
 
 -- | 'MonadError' analogue to the 'Control.Exception.try' function.
 --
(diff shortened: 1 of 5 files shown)


Control/Monad/Error/Class.hs
```diff
@@ -206,6 +207,13 @@ instance
     throwError = lift . throwError
     catchError = Accum.liftCatch catchError
 
+instance (MonadError e m, MonadError e n) => MonadError e (Product m n) where
+    throwError e = Pair (throwError e) (throwError e)
+    catchError (Pair ma na) f = Pair (catchError ma (productFst . f)) (catchError na (productSnd. f))
```
sjshuck: Can you please write `catchError` in infix notation, at least on the right-hand side?  It can reduce nested parentheses.  Also the final `.` needs space around it.

LightAndLight: > Can you please write catchError in infix notation, at least on the right-hand side?

When I do this:

```
Pair (ma `catchError` productFst . f) (na `catchError` productSnd . f)
```

I get a precedence parsing error (`catchError` is `infixl 9` and `.` is `infixr 9`).

I could use infix `Pair` to reduce parentheses:

```
catchError ma (productFst . f) `Pair` catchError na (productSnd . f)
```

Thoughts?

LightAndLight: > Also the final . needs space around it.

Fixed in 22c80e7

sjshuck: > Thoughts?

My thoughts are that my cellphone is not the best tool for collaborating on GitHub issues.

Thanks for the contribution!

Control/Monad/State/Class.hs
```diff
@@ -192,3 +193,7 @@ instance MonadState s m => MonadState s (SelectT r m) where
     get = lift get
     put = lift . put
     state = lift . state
+
+instance (MonadState s m, MonadState s n) => MonadState s (Product m n) where
+    get = Pair get get
+    put s = Pair (put s) (put s)
```
mniip: An explicit implementation of `state` would probably be good.

Control/Monad/Writer/Class.hs
```diff
@@ -205,3 +206,8 @@ instance
     tell   = lift . tell
     listen = Accum.liftListen listen
     pass   = Accum.liftPass pass
+
+instance (MonadWriter w m, MonadWriter w n) => MonadWriter w (Product m n) where
+    tell w = Pair (tell w) (tell w)
+    listen (Pair ma na) = Pair (listen ma) (listen na)
+    pass (Pair maf naf) = Pair (pass maf) (pass naf)
```
mniip: An explicit implementation of `writer` would probably be good.

mniip (commented):
A note regarding strictness: I see that all instances are strict in the `Pair` constructor, and I wonder whether that's correct. However I also note that the instances in `base` (particularly `Monad`) are also strict.

PR #179, opened 2026-02-13
Update link for Monad Transformers resource

fishtreesugar:


L0neGamer (approved):
Good to update away a dead link!

As per https://github.com/sdiehl/wiwinwlh, the `github.io` url is the correct one to use.

PR #182, opened 2026-04-11
Overhaul CI

hasufell:
What this PR does:

- test the sdist, not the project!
  - we want to know it can be built from hackage
  - that means we also get rid of `cabal freeze`
  - and inject `--ghc-options='-Werror'` manually
- automatically test all GHCs in the `tested-with` field
  - we use https://github.com/Kleidukos/get-tested
  - due to an outstanding issue we need a minor workaround: https://github.com/Kleidukos/get-tested/issues/93
- upload the `plan.json` files
- automatically test the last 4 LTS stackage releases
  - this code is custom shell, but very simple
  - this should also implicitly test that we're compatible with the shipped boot libraries
- test haddock
- run `cabal check`
- remove caching, we don't really need it
- use `haskell/ghcup-setup` over `haskell-actions/setup`
  - it's simpler and more predictable
  - allows better advanced configuration
- add matrix to mhs job just in case
- update `actions/checkout` to `v6`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,79 +7,192 @@ on:
   push:
     branches: [ "master" ]
 
+defaults:
+  run:
+    shell: bash
+
 jobs:
-  ghc:
+  # this action can extract the Cabal 'tested-with' field
+  tested-with-matrix:
+    name: "Generate matrix from cabal"
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Extract the tested GHC versions
+        id: set-matrix
+        uses: kleidukos/get-tested@v0.1.9.1
+        with:
+          cabal-file: mtl.cabal
+          ubuntu-version: "latest"
+          macos-version: "latest"
+          windows-version: "latest"
+          version: 0.1.9.1
+
+  # workaround for https://github.com/Kleidukos/get-tested/issues/93
+  # we only want the GHC versions, not the whole matrix
+  tested-with-ghcs:
+    name: "Extract GHC versions from tested-with-matrix"
+    runs-on: ubuntu-latest
+    needs: ["tested-with-matrix"]
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+
+    steps:
+    - name: Generate output
+      id: set-matrix
+      run: |
+        set -euo pipefail
+
+        # input might be something like:
+        #   {"include":[{"ghc":"8.10","newest":false,"oldest":true ,"os":"macos-latest"}
+        #              ,{"ghc":"9.2" ,"newest":false,"oldest":false,"os":"macos-latest"}
+        #              ]}
+        # and we want:
+        #   ["8.10","9.2"]
+        ghc_versions=$(echo '${{ needs.tested-with-matrix.outputs.matrix }}' | jq -c -r '[.include[].ghc] | unique')
+
+        echo "${ghc_versions}"
+        echo matrix="${ghc_versions}" >> "$GITHUB_OUTPUT"
+
+  lts-matrix:
+    name: "Generate matrix of last n Stackage LTSes"
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+
+    steps:
+    - name: Generate output
+      id: set-matrix
+      run: |
+        STACK_TEST_NUM=4
+        set -euo pipefail
+
+        # the input is something like this:
+        #   {"lts":"lts-24.36", "lts-0":"lts-0.7", "lts-1":"lts-1.15", "nightly":"nightly-2026-04-10"}
+        # and we want the last n LTS, sorted and deduplicated without nightlies:
+        #   ["lts-0.7", "lts-1.15", "lts-24.36"]
+        lts_versions=$(curl -s https://www.stackage.org/download/snapshots.json | jq -c -r --arg n "$STACK_TEST_NUM" '
+          [.[] | select(startswith("nightly") | not)]
+          | unique
+          | sort_by(split("-")[1] | split(".") | map(tonumber))
+          | .[-(($n | tonumber)):]
+        ')
+        echo "${lts_versions}"
+        echo matrix="${lts_versions}" >> "$GITHUB_OUTPUT"
+
+  cabal:
     name: "GHC ${{ matrix.ghc }} on ${{ matrix.os }}"
     runs-on: ${{ matrix.os }}
+    needs: ["tested-with-ghcs"]
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        ghc: ['8.10', '9.2', '9.8', '9.10', '9.12']
+        ghc: ${{ fromJSON(needs.tested-with-ghcs.outputs.matrix) }}
         exclude:
           - os: macos-latest
             ghc: '8.10'  # ghc-8.10 does not support ARM
       fail-fast: false
     steps:
       - name: Checkout base repo
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
 
       - name: Set up Haskell
         id: setup-haskell
-        uses: haskell-actions/setup@v2
+        uses: haskell/ghcup-setup@v1
         with:
-          ghc-version: ${{ matrix.ghc }}
-          cabal-version: 'latest'
+          ghc: ${{ matrix.ghc }}
+          cabal: 'latest'
 
-      - name: Configure
-        run: cabal new-configure
+      - name: Cabal update
+        run: |
+          cabal update
 
-      - name: Freeze
-        run: cabal freeze
-
-      - name: Cache
-        uses: actions/cache@v4
-        with:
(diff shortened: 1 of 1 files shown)


sjshuck:
Reviewing this now @hasufell, thanks.
> due to an outstanding issue we need a minor workaround

Is this a blocker?

hasufell:
No, the workaround is in the workflow file. I documented it there.

.github/workflows/ci.yml
```diff
@@ -7,53 +7,184 @@ on:
...
+      run: |
+        set -euo pipefail
+
+        # input might be something like:
+        #   {"include":[{"ghc":"8.10","newest":false,"oldest":true ,"os":"macos-latest"}
+        #              ,{"ghc":"9.2" ,"newest":false,"oldest":false,"os":"macos-latest"}
+        #              ]}
+        # and we want:
+        #   ["8.10","9.2"]
+        ghc_versions=$(echo '${{ needs.tested-with-matrix.outputs.matrix }}' | jq -c -r '[.include[].ghc] | unique')
```
sjshuck, line 53: I know the GHC version number is never going to have a space in it, but maybe quote the command substitution

.github/workflows/ci.yml
```diff
@@ -7,53 +7,184 @@ on:
...
+    name: "Generate matrix of last n Stackage LTSes"
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+
+    steps:
+    - name: Generate output
+      id: set-matrix
+      run: |
+        : "${STACK_TEST_NUM:=4}"
```
sjshuck: Can you write this as boring `if` statement?

.github/workflows/ci.yml
hasufell: You don't need to quote in this case. It works even of there are spaces.

.github/workflows/ci.yml
```diff
@@ -7,53 +7,184 @@ on:
...
+        set -euo pipefail
+
+        # the input is something like this:
+        #   {"lts":"lts-24.36", "lts-0":"lts-0.7", "lts-1":"lts-1.15", "nightly":"nightly-2026-04-10"}
+        # and we want the last n LTS, sorted and deduplicated without nightlies:
+        #   ["lts-0.7", "lts-1.15", "lts-24.36"]
+        lts_versions=$(curl -s https://www.stackage.org/download/snapshots.json | jq -c -r --arg n "$STACK_TEST_NUM" '
```
sjshuck, line 75: Quote command substitution

.github/workflows/ci.yml
```diff
@@ -7,53 +7,184 @@ on:
...
+          [.[] | select(startswith("nightly") | not)]
+          | unique
+          | sort_by(split("-")[1] | split(".") | map(tonumber))
+          | .[-(($n | tonumber)):]
```
sjshuck, line 79: Double parens needed?

.github/workflows/ci.yml
hasufell: `${var:=default}` is a very common idiom that sets var to default if it's unset or empty. The colon is a shell built-in that just makes sure the substitution isn't executed as a command: it does nothing.

.github/workflows/ci.yml
```diff
@@ -7,53 +7,184 @@ on:
...
+        ')
+        echo "${lts_versions}"
+        echo matrix="${lts_versions}" >> "$GITHUB_OUTPUT"
+      env:
+        # The default is 4.
+        # We can set a repository variable that overwrites this,
```
sjshuck: I like the functionality, but it is magical insofar as you don't really see it in the action log or the CI YAML.  And it's not needed yet, and now is the time we can just decide on our policy.  My preference would be to hard-code it to 4 in the CI YAML.

.github/workflows/ci.yml
```diff
@@ -7,53 +7,184 @@ on:
...
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
 
       - name: Set up Haskell
         id: setup-haskell
-        uses: haskell-actions/setup@v2
+        uses: haskell/ghcup-setup@v1
+        with:
+          ghc: ${{ matrix.ghc }}
+          cabal: 'latest'
```
sjshuck, line 105: Is there a way of changing this to the Cabal version that's associated with whichever GHC?  It may be a non-issue&mdash;I personally always keep Cabal at latest.

.github/workflows/ci.yml
```diff
@@ -7,53 +7,184 @@ on:
...
-          key: ${{ runner.os }}-${{ matrix.ghc }}-${{ hashFiles('cabal.project.freeze') }}
-          restore-keys: ${{ runner.os }}-${{ matrix.ghc }}-
+          stack: 'latest'
+          stack-hook: false
 
       - name: Build
-        run: cabal build
+        run: |
+          # just in case
+          rm -f stack.yaml stack.yaml.lock
```
sjshuck: This is a good point.  I almost wonder if we should have it hard error if found.

.github/workflows/ci.yml
hasufell: `cabal-install` is generally backwards compatible

.github/workflows/ci.yml
```diff
@@ -74,7 +205,7 @@ jobs:
           mcabal install
```
sjshuck: `-r`
See https://github.com/haskell/transformers/pull/116/changes

.github/workflows/ci.yml
```diff
@@ -7,53 +7,184 @@ on:
...
+          stack init --resolver=${{ matrix.lts }}
+          # build
+          stack build
 
   mhs:
+    name: "MicroHS ${{ matrix.mhs }}"
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        mhs: ["0.14.15.0"]
```
sjshuck: You can bump this version per https://github.com/haskell/transformers/pull/116/changes I think

.github/workflows/ci.yml
```diff
@@ -62,7 +193,7 @@ jobs:
           make minstall
 
       - name: checkout transformers repo
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
         with:
           repository: augustss/transformers
```
sjshuck: I think we can change this to `haskell/transformers` now

.github/workflows/ci.yml
sjshuck: Indeed!

.github/workflows/ci.yml
sjshuck: Yes, I understood it, but another I was pairing with did not.  Do you have a strong preference for this way?

.github/workflows/ci.yml
sjshuck: EDIT Not needed

.github/workflows/ci.yml
sjshuck: Agreed.  If there's an easy way to set this to the associated version, I prefer that, since updating the cabal file with newer features will break anyone stuck on older versions (they are already stuck on older GHCs...).  If there's no easy way, it's fine.

.github/workflows/ci.yml
```diff
@@ -62,7 +193,7 @@ jobs:
           make minstall
```
konsumlamm, line 187:
```suggestion
- name: make and install mhs
run: |
         cd mhs
         make minstall
         echo "$HOME/.mcabal/bin" >> "$GITHUB_PATH"
```
This adds `~/.mcabal/bin` (where the `mcabal` and `mhs` binaries are located) to the path, so that it's available in later steps.

.github/workflows/ci.yml
konsumlamm: This whole step isn't needed anymore, since the Hackage version of `transformers` compiles with MicroHs now.

.github/workflows/ci.yml
konsumlamm: This step isn't needed either anymore.

.github/workflows/ci.yml
konsumlamm:
```suggestion
run: |
         cd mtl
         mcabal -r install
```
Setting the path isn't necessary anymore with the above change. The `-r` flag recursively installs all dependencies (the Hackage versions all work in this case).

.github/workflows/ci.yml
sjshuck: How do you suggest we tell Cabal or Stack to pull in `master` of transformers and not the boot version?
EDIT You said the Hackage version.  I don't recall anything changing there; did a MicroHs change fix that?

konsumlamm: Possibly, I don't remember why it didn't work before.

.github/workflows/ci.yml
hasufell: Sorry, I don't have such a list and I don't feel like implementing that. But feel free to adjust this PR.

.github/workflows/ci.yml
hasufell: I'm not sure I follow. The whole thing is already in single parens.

sjshuck (approved):

konsumlamm (approved):

hasufell:
Cheers

PR #183, opened 2026-06-08
Remove cpp from monadcont

sjshuck:


L0neGamer (commented):
Good to see that we don't need the kind signature, and I appreciate the test assertion.

I can't see that the CI has run, can you verify that it is meant to?

EDIT: I was too fast, it's going now.

Kinda crazy that there are no tests here...

L0neGamer (approved):

